PR #13, opened 2019-11-06
Updated eigenvalue solver for complex matrices

MarkGillespie:
The current k-eigenvalue solver does not work on complex matrices. For complex vectors, `x.transpose() * v` does not give the dot product. I switched to Eigen's `dot` function, which should work in both the real and complex case. Also for projections onto complex vectors, you need to write the dot product in a particular order - `v.dot(x) != x.dot(v)`. So I switched the order of `v` and `x` in the projection.


nmwsharp:
Awesome, thank you!

PR #17, opened 2020-02-25
Added STL parser

MarkGillespie:
I added parsers for ASCII and binary STL files. I also added a function to merge vertices with the same positions, which lets you build halfedge meshes out of the STL models.

`mergeIdenticalVertices()` assumes that the vertices have exactly the same coordinates in floating point - I tried a version that merged vertices up to a small tolerance, but that was significantly slower, and this seems to work fine on the few Thingi10k models that I tried.

### DIFF
--- a/include/geometrycentral/surface/polygon_soup_mesh.h
+++ b/include/geometrycentral/surface/polygon_soup_mesh.h
@@ -3,18 +3,21 @@
 #include <cstdlib>
 #include <fstream>
 #include <memory>
+#include <regex>
 #include <stdexcept>
+#include <unordered_map>
 #include <vector>
 
-#include "geometrycentral/utilities/vector3.h"
 #include <geometrycentral/utilities/utilities.h>
+#include <geometrycentral/utilities/vector2.h>
+#include <geometrycentral/utilities/vector3.h>
 
 namespace geometrycentral {
 
 class PolygonSoupMesh {
 public:
   PolygonSoupMesh();
-  PolygonSoupMesh(std::string meshFilename);
+  PolygonSoupMesh(std::string meshFilename, std::string type = "");
   PolygonSoupMesh(const std::vector<std::vector<size_t>>& polygons_, const std::vector<Vector3>& vertexCoordinates_);
 
   // Mutate this mesh and by naively triangulating polygons
@@ -24,8 +27,16 @@ public:
   std::vector<std::vector<size_t>> polygons;
   std::vector<Vector3> vertexCoordinates;
 
+  // Mutate this mesh by merging vertices with identical floating point positions.
+  // Useful for loading .stl files, which don't contain information about which
+  // triangle corners meet at vertices.
+  void mergeIdenticalVertices();
+
 private:
-  void readMeshFromFile(std::string filename);
+  void readMeshFromObjFile(std::string filename);
+  void readMeshFromStlFile(std::string filename);
+  void readMeshFromAsciiStlFile(std::ifstream& in);
+  void readMeshFromBinaryStlFile(std::ifstream in);
 };
 
 } // namespace geometrycentral
--- a/include/geometrycentral/surface/polygon_soup_mesh.h
+++ b/include/geometrycentral/surface/polygon_soup_mesh.h
@@ -3,7 +3,6 @@
 #include <cstdlib>
 #include <fstream>
 #include <memory>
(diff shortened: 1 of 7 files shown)


include/geometrycentral/surface/polygon_soup_mesh.h
```diff
@@ -3,18 +3,21 @@
 #include <cstdlib>
 #include <fstream>
 #include <memory>
+#include <regex>
```
nmwsharp: looks like this isn't used, right? can we remove it?

src/surface/polygon_soup_mesh.cpp
```diff
@@ -10,7 +10,36 @@ namespace geometrycentral {
...
+  // Attempt to detect filename
+  bool typeGiven = type != "";
+  std::string::size_type sepInd = meshFilename.rfind('.');
+  if (!typeGiven) {
+    if (sepInd != std::string::npos) {
+      std::string extension;
+      extension = meshFilename.substr(sepInd + 1);
+
+      // Convert to all lowercase
+      std::transform(extension.begin(), extension.end(), extension.begin(), ::tolower);
```
nmwsharp: Why doesn't this need to be `std::tolower`? Do we have a `using namespace std` somewhere or something? How do namespaces work?!

src/surface/polygon_soup_mesh.cpp
```diff
@@ -120,6 +148,177 @@ void PolygonSoupMesh::readMeshFromFile(std::string filename) {
...
+  size_t lineNum = 1;
+
+  auto assertToken = [&](const std::string& expected) {
+    std::string token;
+    ss >> token;
+    if (token != expected) {
+      std::cerr << "Error on line " << lineNum << ". Expected \"" << expected << "\" but token \"" << token << "\""
+                << std::endl;
+      std::cerr << "Full line: \"" << line << "\"" << std::endl;
+      exit(1);
```
nmwsharp: Should probably `throw std::runtime_error()` here, rather than `exit(1)`. A user might want to catch an exception if loading fails.

Also, `std::cerr` is fine, but I tend to prefer using `std::cout` everywhere. The reason why is that when we intermingle `std::cout` and `std::cerr`, the print statements can end out of order in the terminal, which gets confusing. Using only `std::cout` keeps a global ordering.

src/surface/polygon_soup_mesh.cpp
```diff
@@ -120,6 +148,177 @@ void PolygonSoupMesh::readMeshFromFile(std::string filename) {
...
+      canonicalIndex[v] = vecIndex;
+      compressVertex.push_back(vecIndex);
+    } else {
+      size_t vecIndex = it->second;
+      compressVertex.push_back(vecIndex);
+    }
+  }
+
+  vertexCoordinates = std::move(compressedPositions);
+  compressedPositions.clear();
```
nmwsharp: Calling a method on `compressedPositions` after moving from it is undefined behavior ("use after move"), right? I think this line should be omitted.

include/geometrycentral/surface/polygon_soup_mesh.h
MarkGillespie: Yeah, this was left over from a previous version of the code.

src/surface/polygon_soup_mesh.cpp
MarkGillespie: I really don't understand what's going on here, but it only works as `::tolower`, not `std::tolower`. I grabbed this code from `meshio.cpp`, which also doesn't have an `std`. I guess I just don't understand namespaces.

src/surface/polygon_soup_mesh.cpp
MarkGillespie: I updated this to print everything in the `std::runtime_error`.

src/surface/polygon_soup_mesh.cpp
MarkGillespie: You're right. I misunderstood an answer on StackExchange. I've deleted the `clear()`.

src/surface/polygon_soup_mesh.cpp
nmwsharp: Oh good point. I guess I don't understand either. This can remain a mystery for now!

nmwsharp:
Thank you!

PR #19, opened 2020-03-30
Removing absolute library paths for dependencies

ctlee:
In some configurations, libraries such as `-lm` and `-ldl` are referenced by SparseSuite. Appending a path to these flags will produce a linking error later on.

Implemented the use of `INTERFACE` libraries. Newer versions of CMake can support inheritance of includes and linked libraries. Treating everything as a target helps with this.

**Todo:**
- [ ] Test PR on Windows. I unfortunately don't have a suitable Windows developing environment setup right now.

**Changed:**
- Added [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) function from Kitware/CMake which is slightly more sophisticated than `download_project`. Namely it allows for auto population of content with several workflows. It's bundled with CMake >= 3.11 but I noticed that the current restriction is >= 3.8.
- Eigen is now included as an `INTERFACE` target. 
- `Happily`, `nanort`, and `nanoflann` have also been made `INTERFACE` libraries. Similar to [Eigen](https://eigen.tuxfamily.org/dox/TopicCMakeGuide.html) it's possible to automatically export a `target` for each library
- Removed conversions to absolute paths
- SparseSuite libs and includes are now specific to target `geometry-central`. It's a pain to create interface libraries for each component of SparseSuite. Also order of import can matter which complicates things... Also the compile_definition `-DGC_HAVE_SUITESPARSE` is also moved to be target specific.


nmwsharp:
Thanks for contributing! This seems great.

I'm a bit of a cmake amateur, so I want to be sure I understand what's going on. The changes for end users will be:
- Resolving bugs like the absolute path `-lm` issue you mention
- Following more modern cmake practices :+1: 
- The cmake version requirement **does not** change
- Recipes like `target_link_libraries(your-project-target geometry-central)` are unchanged
- Eigen is still resolved [as described here](http://geometry-central.net/build/dependencies/#eigen)

Sound right? Anything else?

I can take care of testing on Windows; I keep a machine around for that purpose. Actually, I should be able to just upgrade the CI to include Windows; I think I've gotten the hang of Windows builds on travis.

ctlee:
Your notes are correct with a few modifications/notes:

> The cmake version requirement **does not** change

There's a workaround to use newer CMake features. `FetchContent` is introduced in 3.11 and `FetchContent_MakeAvailable` in 3.14. I added a local copy of `FetchContent` borrowed from CMake. 

Commit 3747a9f adds in some checks of the CMake version to load a local copy of these functions. We could always load the local copy (`FetchContentLocal`) but then newer features in later CMake versions won't be available.

In the future, when CMake >= 3.14 is standard the local copy of `FetchContent` will be obsolete.

> Recipes like `target_link_libraries(your-project-target geometry-central)` are unchanged

Yes this should be the case. This PR could benefit from some additional care in specifying if linked libraries are in the public API of geometry-central. 

At this point, libraries in `GC_DEP_LIBS` (Eigen, happly, nanort, and nanoflann) are `PUBLIC`. Meaning that downstream targets linking to `geometry-central` will be passed the headers of these libraries as well. [This](https://cmake.org/pipermail/cmake/2016-May/063400.html) discussion was helpful to me.

> Resolving Eigen...

FetchContent downloads a copy of Eigen to the `SOURCE_DIR` (e.g., `build/deps/eigen-src`) which is slightly different than the description.
```
SOURCE_DIR      "${CMAKE_CURRENT_BINARY_DIR}/eigen-src"
BINARY_DIR        "${CMAKE_CURRENT_BINARY_DIR}/eigen-build"
```

- With respect to building the library on Windows. 

I'm definitely not an expert... I think one has to explicitly export things to the library using `__declspec(dllexport)` and other techniques. When I tried building on Windows I don't think anything was exported.

There's a CMake cheat to export everything: `set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)`

I'm not sure how you're compiling and linking in Windows currently, but this may be something to add.

- With respect to installing Suitesparse on Windows. Does Anaconda install a suitable version? CMake found SuiteSparse from my Anaconda install which is how I discovered the `-lm` issue.

test/CMakeLists.txt
```diff
@@ -54,35 +56,27 @@ else()
...
-  WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/googletest-download )
-if(result)
-  message(FATAL_ERROR "CMake step for googletest failed: ${result}")
-endif()
-execute_process(COMMAND ${CMAKE_COMMAND} --build .
-  RESULT_VARIABLE result
-  WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/googletest-download )
-if(result)
-  message(FATAL_ERROR "Build step for googletest failed: ${result}")
+if(${CMAKE_VERSION} VERSION_LESS 3.14)
```
nmwsharp, line 61: I notice that this tests against version 3.14 here, but 3.11 in `deps/CMakeLists.txt`. Is the difference intentional?

nmwsharp:
OK! All those responses make sense to me, thanks for the explanations. I appreciate the extra effort to preserve the CMake version requirement!

I didn't think about testing this against Suitesparse on Windows... I don't know of any good solution for that, and have not tried the conda package. Once upon a time I manually compiled/installed on Windows, and it was very unpleasant. We've always said Suitesparse/Windows is a use-at-your-own-peril combination for geometry-central, so I'm okay with not testing it.

As to other Windows best-practices, let's not make any changes now. They can go in a separate PR, if/when someone on Windows complains :smile: 

This is still a bit out of my expertise, but everything looks good to me (modulo one small inline comment I just made).

Is this done as far as you're concerned? If so I'll do a little manual testing and then we can merge it. Thanks for all the work!

test/CMakeLists.txt
ctlee: **This difference is intentional!** 
One of my frustrations with CMake is that the API is a moving target. Theres a call to `FetchContent_MakeAvailable()` which was introduced in 3.14. The file/module defining these other calls are copied verbatim from some future CMake so we can use this function. When CMake 3.14 is commonplace then it would make sense to use FetchContent definitions from the CMake install.

**Why does FetchContent for Eigen only require 3.11?** 
That's when FetchContent was introduced in general. Here since `FetchContent_MakeAvailable` will build Eigen tests and other things which are undesirable, we're only using the `Populate` and other calls which defined in the original FetchContent module. Actually the following code to populate etc is very similar to that in the contents of `MakeAvailable`. We just remove the part to recurse into the subdirectory. We substitute in some logic to check the Eigen headers and define the relevant target which is the minimal thing to do.

There is one potential motivation to recurse into the Eigen subdirectory. Eigen CMakeLists.txt defines a bunch of compiler tests and flag definitions which may improve the performance of a subsequent build. I haven't looked closely at this and can't comment at this point on if it will actually matter or not.

ctlee:
This PR is complete as far as I'm concerned. Thanks for developing this helpful library.

nmwsharp:
Merged! Thank you for the great PR!

nmwsharp:
I didn't realize until after merging that this line:

https://github.com/nmwsharp/geometry-central/pull/19/files#diff-95e351a3805a1dafa85bf20b81d086e6R110

Is potentially problematic. It forces a shared build of geometry-central, even if the user might be trying to do a static build of their library. This caused problems for existing CMake setups.

For now I immediately pushed another commit which removes `shared`. If there's some important reason for it to be there let me know!

nmwsharp:
(the line is the `shared` in https://github.com/ctlee/geometry-central/blob/cce5d9eee1c12dc3c073750c1385baa54f5ed6c5/src/CMakeLists.txt#L110)

ctlee:
You're right in removing the `shared` keyword. CMake will default to the value of `BUILD_SHARED_LIBS` to set whether a shared or static library is built which is the desired behavior.

PR #24, opened 2020-04-21
Unset EIGEN3_INCLUDE_DIR on failure

ctlee:
Under certain awful system configuration conditions... it's apparently possible for `find_path()` to set a bogus Eigen location in `EIGEN3_INCLUDE_DIR`. Subsequent calls to `find_path()` will default to the previously found location even though its wrong (c.f. https://cmake.org/cmake/help/latest/command/find_path.html second to last paragraph).

Also `find_path(<var> ...)` expects a value of the form `<var>-NOTFOUND` or if the variable is uninitialized. The empty string value, for example, appears to also satisfy the cached value check.

This PR explicitly sets `EIGEN3_INCLUDE_DIR` to value `EIGEN3_INCLUDE_DIR-NOTFOUND` to avoid using an incorrect cached path in the future.


nmwsharp:
Seems reasonable, thanks!

PR #27, opened 2020-06-27
Fixed some bugs in the docs

MarkGillespie:
I wrote a python script to run all the snippets of code in the docs. The script horrible and probably shouldn't be merged into geometry-central in its current state, but I found these bugs with it.

If you're curious, the script is here: https://github.com/MarkGillespie/geometry-central/blob/TestDocs/docs/test.py. I'd be happy to talk about it some time.


nmwsharp:
This is amazing! I'm totally interested, both to hear how you did this, and to make use of this script. We should definitely talk.

I'm prepping a whole bunch of updates to the docs for a new version; perhaps we can re-run this script before merging the new version? I'll ping you separately.

PR #28, opened 2020-07-02
Added direction field docs

MarkGillespie:
Here are the docs that I wrote for `direction_fields.h`. At the moment, it looks like only `computeSmoothestVertexDirectionField` is actually implemented, although there's old code for some of the other functions.

I made 2 minor changes to the function signatures in these docs:
1. I added a flag to `computeSmoothestBoundaryAlignedDirectionField` to specify whether the field should be tangent to the boundary or normal to the boundary.
1. I changed the default value of `nSym` to 2 in `computeCurvatureAlignedDirectionField` since it looked like this function only works for `nSym = 2` or `nSym = 4`


dcoeurjo:
Just jumping into the discussion concerning the deadcode in `direction_fields.h`: @nmwsharp would that just be some API updates with the new `SurfaceMesh` or there are more deep changes required?

nmwsharp:
The direction fields code is 90% of the way there to being ready, just need to find the time to polish it and test a little :)

I think there are a few updates to `direction_fields.h` beyond simply the new `SurfaceMesh`, if I recall some of the boundary conditions aren't implemented properly with regards to symmetry. However those should also be straightforward fixes!

I can bump this towards the top of my priority list, since it sounds popular. (And Mark has already written excellent docs!)

DavidJourdan:
Hi, sorry if this is not the right place to ask but for a project I needed face-based direction fields which are commented out at the moment, so I updated the `computeSmoothestFaceDirectionField` function in my own fork, is it okay if I send a pull request for that?

nmwsharp:
Merged! Much later 😛 

Thanks again for writing these! I also added a few screenshots, and docs of the newly implemented face direction field functions.

@dcoeurjo I believe everything is cleaned up with direction fields now. Please post if you still see missing things!

dcoeurjo:
👏 thx

PR #30, opened 2020-07-03
bugfix GCC8.2 -Werror=catch-value=

MrBurnst:
I had a compile error with GCC8.2 with -Werror and the exceptions ([https://www.nsnam.org/bugzilla/show_bug.cgi?id=2917](https://www.nsnam.org/bugzilla/show_bug.cgi?id=2917))


nmwsharp:
ah good catch. Thank you!

PR #44, opened 2020-09-10
signpost_intrinsic_triangulation: implement equivalentPointOn{Intrinsic,Input}

kenshi84:
I played with the signpost intrinsic triangulation demo and realized that `equivalentPointOn{Intrinsic,Input}` are left unimplemented, so here they are. I needed them for testing my own little research idea that turned out to be doomed, but hopefully someone will find this code useful in the future.

This implementation has a robustness/accuracy issue despite my attempt, as evidenced in the test. In some rare but non-negligible cases, the positional error after going to/from inputMesh and intrinsicMesh gets quite large (more than 50% of the bounding box diagonal). Hopefully someone can fix this bug in the future :)

P.S.
Thank you so much for maintaining this awesome library!

### DIFF
--- a/include/geometrycentral/surface/signpost_intrinsic_triangulation.h
+++ b/include/geometrycentral/surface/signpost_intrinsic_triangulation.h
@@ -64,10 +64,10 @@ public:
   // ======================================================
 
   // Given a point on the input triangulation, returns the corresponding point on the intrinsic triangulation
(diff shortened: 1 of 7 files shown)


nmwsharp:
Thanks for sharing this, it looks awesome! I really appreciated it.

I haven't gotten a chance to look deeply yet, but it's on my agenda! Hopefully I can get this merged soon.

PR #45, opened 2020-10-16
Add Missing Direction fields

DavidJourdan:
For a project I needed to play with direction fields, so I brought back all the commented out functions by trying to keep them as close as possible to the original implementations while using the new API.

I tested all the functions and they seem to be working, although when comparing to [this implementation](https://github.com/GeometryCollective/fieldgen) the fields seem much more smoothed out (there's less singularities), is it because the mass and energy matrices are slightly different than in the paper?

Anyways, thanks for this really cool library!

Edit: I realized that the connection Laplacian computed in IntrinsicGeometryInterface is only valid for nSym = 1, so I added the correct functions for computing these laplacians in an anonymous namespace


nmwsharp:
This looks wonderful, thanks so much for taking the time to contribute!

I don't immediately know of any differences that would change results vs. the fieldgen implementation, but I can take a quick look to see if I spot anything.

We should definitely get this merged! I notice you've been making a few more updates, let me know when it seems final on your end and I'll do a pass through.

DavidJourdan:
Yes I think I'm done with the updates, please tell me if you need me to do any more modifications to get this merged. 

The only thing which is unclear to me is the difference with fieldgen. From what I understood reading Knoppel et al.'s paper, the vertex connection laplacian and galerkin mass matrix seem to be approximations of the energy and mass matrices (eq. 17-18) in the limit of no curvature, but maybe I'm missing something? The implementation as it is seems to be working just fine, so I'm not too worried about it anyway

nmwsharp:
Ok---I pulled this down and it mostly looks great!

One important question: I don't really understand what's being done here for the boundary-aligned case. It looks like the boundary term is just being used as the RHS in a Poisson equation, but it should instead be Dirichlet boundary conditions, right? In practice, I noticed this because the results look a bit off on most test inputs with nsym > 1.


 > RE: differences vs. fieldgen

I don't believe Knoppel et. al. has anything to say about discretization for face-based fields. And the "unit-weight" Laplacian used here with w=1 doesn't really have any physical meaning, so it won't converge or yield the same results even under refinement (though I'm still fine with using w=1, since there's no other obvious choice). As you mentioned, for vertex-based fields the connection Laplacian we build is not the same as proper finite element discretization given in Eqn 18 of Knoppel et al, and thus may yield different results. I don't know whether one could argue for convergence between the definitions in some limit.

DavidJourdan:
> It looks like the boundary term is just being used as the RHS in a Poisson equation, but it should instead be Dirichlet  boundary conditions, right?

Oh I think I see what you mean, I'll try fixing that quickly.  

I'm fine w=1 too, I was just curious about the differences in the discretisation of the Laplacian for vertex-based fields and wether or not this could explain the differences I noticed between the fieldgen and geometry-central implementations. But as you said, these differences are probably due to  an error in the way I handled the boundary conditions.

DavidJourdan:
I fixed the boundary aligned fields, and took the opportunity to also bring back the index computation for finding singularities.

I think there is still an issue with the curvature fields, but could not figure out what was wrong. Here is a comparison between fieldgen and geometry-central: 

Screenshot 2020-10-24 at 17 35 43
Screenshot 2020-10-24 at 17 35 23

The curvature-aligned field in geometry-central doesn't have any singularity and looks almost like the smoothest direction field

DavidJourdan:
Never mind I fixed the bug, it was an error in the computation of principal curvature directions.

I think everything works as expected now, so hopefully this is my last commit on this pull request :)

Screenshot 2020-10-25 at 11 14 59

nmwsharp:
Awesome! I just pulled it down and everything looks great :)  I'm going to add some docs with #28 so folks know that this is there.

Thanks for the contribution!

nmwsharp:
also, do you have a twitter handle or anything? I'm gonna tweet at some point to advertise the new direction field improvements and would be happy to give you a shout out.

nmwsharp:
By the way, I also removed this comment about extrinsic dirichlet energy

https://github.com/DavidJourdan/geometry-central/blob/direction-fields/include/geometrycentral/surface/direction_fields.h#L28

because it didn't seem to be true in the implementation. Let me know if that doesn't sound right!

DavidJourdan:
Yes, that sounds right. I believe it was a comment that was already there before

My twitter handle is https://twitter.com/dJourdan_ ;)

PR #46, opened 2020-10-26
Implement edge collapse and edge removal

MarkGillespie:
I got edge collapse/removal working, at least on meshes without boundary. I haven't really tested this at all, but it seems to be working for me so I figured I would submit a pull request in case it's helpful for anyone else.


nmwsharp:
This is at least partially subsumed by #52. Are there parts of it that are still worth pursuing?

By the way, my apologies for not coordinating work between this and #52! Ideally some effort could have been saved.

PR #49, opened 2020-11-13
Fix triangulate to return list of new faces

MarkGillespie:
Currently it returns the old face n times.

The halfedge returned by `connectVertices` is the halfedge in the old face that is adjacent to the new face. So `connectHe.twin().face()` is guaranteed to be in the newly-created face.


nmwsharp:
awesome, thank you!

nmwsharp:
Huh, looks like the "squash and merge" option stuck my name on the commit instead of yours... Sorry about that!

One day I'll figure out the right policy for which merge option to use when...

PR #52, opened 2020-12-02
Triangular edge collapse

icethrush:
This edge collapse function handles numerous corner cases, can handle meshes with boundary, and has been tested on a variety of situations (though certainly not all possible situations). It seems like the most robust implementation we have right now, and we're currently using it as part of a dynamic meshing routine in our project.


src/surface/surface_mesh.cpp
```diff
@@ -844,6 +844,11 @@ bool SurfaceMesh::flip(Edge eFlip) {
   Vertex vb = hb1.vertex();
   Vertex vc = ha3.vertex();
   Vertex vd = hb3.vertex();
+  // if flipped, make sure it is not a duplicate
+  for(Vertex v : vc.adjacentVertices()){
+    if(v == vd) return false;
+  }
+
```
nmwsharp, line 851: I totally understand how you would probably want to avoid self-edges. But there are some situations where they are desired! In particular, I have a personal vested interest in this change because changing the default behavior like this would break mountains of my intrinsic triangulation code :) 

Can we at least guard this check behind a flag like `flip(preventSelfEdge=false)`?

src/surface/manifold_surface_mesh.cpp
```diff
@@ -1037,6 +1037,242 @@ Vertex ManifoldSurfaceMesh::insertVertex(Face fIn) {
...
+
+      v = vB;
+      return true;
+    }
+
+    else {
+      // return heA0.vertex();
+      throw std::runtime_error("what is this vertex " + std::to_string(heA0.vertex()) + " with degree " +
+                               std::to_string(heA0.vertex().degree()) + " on edge " + std::to_string(heA0.edge()) +
+                               "?\n");
```
nmwsharp, line 1268: could this please be changed to have a simple informative message in the error?

include/geometrycentral/surface/manifold_surface_mesh.h
```diff
@@ -65,12 +64,11 @@ class ManifoldSurfaceMesh : public SurfaceMesh {
   // Collapse an edge. Returns the vertex adjacent to that edge which still exists. Returns Vertex() if not
   // collapsible. Assumes triangular simplicial complex as input (at least in neighborhood of collapse).
   Vertex collapseEdge(Edge e);
+  
+  bool collapseEdgeTriangular(Edge e, Vertex &v);
```
nmwsharp, line 68: Could this be changed to have a signature like:
```
Vertex collapseEdgeTriangular(Edge e);
```
and return the null `Vertex()` when impossible? This call style is used in other mesh mutation operations so it would be nice to be consistent.

Also, the plain old `collapseEdge()` right above also assumes triangular... should both still exist? Is that old `collapseEdge()` obsolete now?

nmwsharp:
Thanks! Added a few comments inline.

I didn't really have time to check the logic, but I'm happy to trust it :) At the very least, I easily believe this is better than what is already there!

If you have time, unit tests and docs updates would be amazing. Though if not thats fine, I can add them later!

src/surface/surface_mesh.cpp
icethrush: Yep, added the optional argument `bool preventSelfEdges = false` to `flip`.

src/surface/manifold_surface_mesh.cpp
icethrush: Yes -- I believe this is the case when an interior vertex has degree < 3, so I've just added a message saying that this case isn't supported for collapses.

include/geometrycentral/surface/manifold_surface_mesh.h
icethrush: Done. Also I think, if the old edge collapse also assumes triangular, then it's obsoleted by this one. Assuming it isn't referenced anywhere else in the code, it should be safe to delete.

icethrush:
This should address the comments. I've also updated the docs to mention `collapseEdgeTriangular` and the new optional argument to `flip`. Unfortunately we don't have any specific unit tests set up right now (we've mostly just been stress testing it by running tons of collapses on large examples), but we can sort those out later.

nmwsharp:
Ok! I pushed a few changes: 

- After thinking about it some more, it really does make more sense for `edgeFlip()` to prevent self-edges by default, since that is what most would expect. I'm the weird one who wants self-edge edges :) I changed that to the default, and I'll go update my code where needed.

- I added a few very simple unit tests, since these things are so easy to break if we ever mess with the data structure internals

- added an explicit check for the triangular condition, so we give a more sane error in that case

- stripped out the old (dead) collapse function

Merging now. Thank you!

PR #53, opened 2020-12-07
Add method for loading UVs with ManifoldSurfaceMesh

keenancrane:
This PR adds a new method `readParameterizedManifoldSurfaceMesh()` that loads an OBJ file with UV coordinates directly into a `ManifoldSurfaceMesh` and associated `CornerData`.  It does not break compatibility with any existing methods for reading data; just adds one that loads UVs.

To implement it, I generalized the method `makeManifoldSurfaceMeshAndGeometry` to take a list of UV coordinates (with the same format produced by `SimplePolygonMesh`).  This method is the "general purpose" method, and returns a list that always includes UVs.  Other wrapper methods take a subset of the return values and return them.

Note that, internally, the mechanism for returning a subset of the return values is a bit ugly, and follows this pattern:

```
   auto lvals = makeManifoldSurfaceMeshAndGeometry(polygons, {}, vertexPositions, {});

   return std::tuple<std::unique_ptr<ManifoldSurfaceMesh>,
          std::unique_ptr<VertexPositionGeometry>>
   ( std::move(std::get<0>( lvals )), // mesh
     std::move(std::get<1>( lvals )) ); // geometry
```

However, this doesn't affect anything on the user end.

### DIFF
--- a/docs/docs/surface/utilities/io.md
+++ b/docs/docs/surface/utilities/io.md
@@ -101,9 +101,17 @@ std::tie(mesh2, geometry2) = readSurfaceMesh("spot_messy.obj");
     When reading from a stream, the type _must_ be specified and cannot be automatically inferred.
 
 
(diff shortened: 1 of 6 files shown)


nmwsharp:
Thanks! I went ahead and pushed directly to the branch to add the nonmanifold variants.

PR #54, opened 2020-12-08
Add vertex scalar curvatures

keenancrane:
GC previously only provided Gaussian curvature at vertices.  This PR adds implementation for a number of other common curvatures namely:

- mean curvature
- minimum principal curvature
- maximum principal curvature

It also adds "immediate" implementations of many existing quantities, namely:

- Gaussian curvature
- dihedral angles
- vertex dual areas

Implementations are provided for both intrinsic/edge length and extrinsic/vertex position geometries.

Note that as with previous immediate-mode implementations, code is duplicated across cached and immediate implementations.  Some care was taken to provide correct values at boundary vertices, but likely more thought is required here.

Docs were updated, including an example of the vertex curvatures.


nmwsharp:
Awesome, thanks!

I haven't tested this myself, but the changes all looks reasonable so we can go ahead and merge.

PR #86, opened 2021-04-17
adding remeshing functions

cbrakensiek:
These new files add four functions: a function that makes all the triangles in the mesh Delaunay, two functions for smoothing vertex positions (circumcenter and Laplacian), and a function for adjusting the mesh so that the edges become a curvature-adaptive target edge length.

### DIFF
--- a/include/geometrycentral/surface/remeshing.h
+++ b/include/geometrycentral/surface/remeshing.h
@@ -1,0 +1,24 @@
+#pragma once
+
+#include "geometrycentral/surface/manifold_surface_mesh.h"
+#include "geometrycentral/surface/vertex_position_geometry.h"
+
+#include <queue>
+
+namespace geometrycentral {
+namespace surface {
+
+// makes all triangles delaunay
+void fixDelaunay(SurfaceMesh& mesh, VertexPositionGeometry& geometry);
+// average positions of vertices based on surrounding vertex positions
+void smoothByLaplacian(SurfaceMesh& mesh, VertexPositionGeometry& geometry);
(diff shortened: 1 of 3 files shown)


nmwsharp:
These all look great, thank you for submitting!

I just gave them a quick read-through and everything looks pretty solid. I'll pull down the code and do some testing soon, probably at the same time as writing docs.

I was thinking it would work well to combine these algorithms with the "mutation manager" functionality currently in this branch https://github.com/nmwsharp/geometry-central/tree/mutation_manager, and do a release together. The mutation manager adds the ability to track data & update arrays during remeshing operations, so it seems like a great synergy with these algorithms. Just let me know if you have other thoughts!

fire:
Is there anything I can improve on this?

mariuszhermansdorfer:
Reviving this thread. In the search for an efficient mesh library to base my Rhino plugin on, I stumbled upon geometry-central. It looks fantastic, but I really need fast remeshing tools. It seems that this PR has not been looked into for more than a year. Andthe only reference to remeshing I can find in the repo is commented out:

https://github.com/nmwsharp/geometry-central/blob/b704f0ef35a94504fca65dbf6b88a8b65ceee4ae/include/geometrycentral/surface/mutation_manager.h#L157

Is this functionality being worked on, or have your plans changed?

MarkGillespie:
Thanks for the reminder, I just added the remeshing code in #137

mariuszhermansdorfer:
Thanks, @MarkGillespie!

PR #87, opened 2021-04-26
Bug..fix?

lakshmipathyarjun6:
It seems like the log map computation did not actually take the orientation of a tangent vector into account. Tried this out with the vector heat demo and this appears to adjust with the correct tangent vector propagation...but not entirely sure why the sign inversion was necessary / why the sweep started from the other side if the real axis (presumably) is oriented in the tv direction.

Adding defaulting to not make other stuff blow up :) 

Feel free to reject if unwarranted!

### DIFF
--- a/include/geometrycentral/surface/vector_heat_method.h
+++ b/include/geometrycentral/surface/vector_heat_method.h
@@ -37,7 +37,7 @@ public:
 
 
   // === The Logarithmic map
-  VertexData<Vector2> computeLogMap(const Vertex& sourceVert, double vertexDistanceShift = 0.);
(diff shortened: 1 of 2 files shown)


nmwsharp:
Thanks for submitting! This seems like a reasonable extension to me.

I think the sign inversion should not actually be needed. You might be getting tricked by the Polyscope visualization in the vector heat demo. There, the "branch cut" discontinuity in the angular color ends up at the `-x` axis, since it's computed via `atan`.

lakshmipathyarjun6:
eh...why it not like 1i? and ah makes sense - will remove the inversion and resubmit

lakshmipathyarjun6:
ah. got it

nmwsharp:
closing old issues

PR #96, opened 2021-10-05
Implement `setMarkedEdges`

MarkGillespie:
Looks like we missed this when implementing the general intrinsic triangulation class. I just copied the function from the old signpost code


nmwsharp:
thanks!

PR #98, opened 2021-11-03
Add algorithm for Geodesic Centroidal Voronoi Tessellations

MarkGillespie:
(This also changes the vector heat method to use a `PositiveDefiniteSolver` for vector diffusion on Delaunay meshes, rather than a `SquareSolver`. This makes the Voronoi tessellation generation significantly faster. Hopefully it doesn't break anything)


nmwsharp:
looks awesome, thank you!

PR #102, opened 2021-12-08
Intrinsic MMP

MarkGillespie:
I ported Kirsanov's implementation of MMP to run in geometry central on intrinsic triangulations. Now it uses geometry central's mesh data structures and does its computations based on edge lengths rather than vertex positions, but all of the algorithm logic (and most of the code) is left exactly the same.

After doing this, I noticed that there is a bunch of commented-out code in `exact_polyhedral_geodesics.h` 😄. What's the status of that stuff? If there are plans for getting it working soon, I'd be happy to make any changes to this code to give a more consistent interface or otherwise delineate the differences more clearly.

Also, I'm not totally sure how licensing is supposed to work here. Kirsanov's code is MIT licensed, so based on a cursory google search it seems like it's legal to include a bunch of the code verbatim. I left his copyright headers at the top of the files along with an added link to the original code. Hopefully that suffices

### DIFF
--- a/docs/docs/surface/algorithms/geodesic_distance.md
+++ b/docs/docs/surface/algorithms/geodesic_distance.md
@@ -4,7 +4,140 @@ Note that distance depends on the _intrinsic_ geometry of a surface (via the `In
 
 ## Polyhedral Distance
 
-TODO document
+These routines use [Danil Kirsanov's implementation](https://code.google.com/archive/p/geodesic/) of the [MMP algorithm for exact polyhedral geodesic distance](https://www.cs.jhu.edu/~misha/ReadingSeminar/Papers/Mitchell87.pdf) to compute geodesic distance (and geodesic paths) along a surface.
+
+This class requires that the input be a manifold triangle mesh, but relies only on the intrinsic geometry (represented by edge lengths).
+
+`#include "geometrycentral/surface/exact_geodesics.h"`
+
+### Single Solves
+
+Geodesic distance from a single source vertex can be computed via the following utility function.
+More general source data, queries in the interior of triangles, and geodesic path extraction can be handled using the stateful version below.
+
+Example
+```cpp
+#include "geometrycentral/surface/exact_geodesics.h"
+#include "geometrycentral/surface/meshio.h"
+
+// Load a mesh
+std::unique_ptr<ManifoldSurfaceMesh> mesh;
+std::unique_ptr<VertexPositionGeometry> geometry;
+std::tie(mesh, geometry) = readManifoldSurfaceMesh(filename);
+
+// Pick a vertex
+Vertex sourceVert = /* some vertex */
+
+// Compute distance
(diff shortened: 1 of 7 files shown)


nmwsharp:
Woah, awesome! This is a great addition.

For licensing, I am also not an expert, but I think that what you've done is reasonable. Since the original code is MIT it's pretty easy to satisfy.

For the old stuff in `exact_polyhedral_geodesics.h`, I _think_ that is an old attempt by Yousuf and I (mainly Yousuf) to implement window-based geodesics that we used for some tests during the signpost project. It mostly worked, but I think I found some bug in it so I commented it out. At this point, it should just be deleted; it almost certainly makes sense to use your new Kirsanov-adapted implementation instead since that is battle-tested.

nmwsharp:
Those Eigen failures in the CI build above are unrelated to your change, right? Perhaps some new Eigen version broke our solver code?

MarkGillespie:
Yeah. I think the issue is that recently Apple's clang added `-Wdeprecated-copy` to `-Wall` or `-Wextra`, so now it complains about Eigen using a deprecated copy (whatever that is). I looked into it briefly because on my newer computer clang won't compile Eigen unless I add `-Wno-deprecated-copy` to the disabled warning list, but on my older computer clang doesn't recognize `-Wno-deprecated-copy`. But I never got around to figuring out how to make the CMake script identify the appleclang version and set the flags accordingly :)

At any rate, the failure seems to be unrelated from my new code

MarkGillespie:
Looking into it a little more, it looks like we could add `-Wno-unknown-warning-option -Wno-deprecated-copy`. Then my older version of clang works fine. But it could be a little risky - it would make it a lot easier to miss typos in the sea of warning flags

nmwsharp:
> clang compile bugs

I think the biggest "bug" here is that we have `-Werror` set on the test build :P I used to be a big `Werror` zealot, but I've come to realize that it's a bad idea on released builds, there's inevitably little errors like this that add unimportant warnings. I've dropped the Werror flag from the test build in b2140a9, and it seems to work now.

Briefly looking at the warning, it really does seem to be some internal Eigen thing, so I'm inclined not to worry about it.

nmwsharp:
Btw, does this code in its current form handle nonmanifold meshes? If so that would be super awesome---I've struggled with distance routines not supporting nonmanifold meshes in the past (regardless we should merge it anyway of course).

If I understand correctly, this algorithm should would pretty much out-of-the box on manifold geometries with only minor tweaks. At nonmanifold edges, we'd need to propagate to _all_ other triangles sharing the edges; at nonmanifold vertices we can just treat them like saddles and spawn new windows.

MarkGillespie:
Currently it only runs on manifold meshes, but I agree that it shouldn't be too hard to run it on nonmanifold meshes! I still don't totally understand how the algorithm works, but your suggestion sounds pretty reasonable

nmwsharp:
Cool :) One day! I'll go ahead and merge. Thanks again!

PR #104, opened 2021-12-18
Nonmanifold MMP

MarkGillespie:
Extend the implementation of MMP to run on nonmanifold `SurfaceMesh` inputs. On manifold portions of the mesh, the algorithm runs unchanged. Whenever it encounters a nonmanifold vertex, it adds that vertex as a pseudosource and propagates a new window to all incident edges. There are probably more efficient strategies which could be used in future, but this seems to work fine for now.

I had to make two minor changes to other parts of the codebase to support this stuff:
1. I tweaked `inFace` and `sharedFace` in `surface_point.ipp`. These functions used to check the neighboring halfedges of an edge using `e.halfedge()` and `e.halfedge().twin()` explicitly---I updated them to loop over `e.adjacentHalfedges()` instead so that they work as expected on nonmanifold edges.
2. I added functions in `SurfaceMesh` and `ManifoldSurfaceMesh` to return a `VertexData<bool>` indicating which vertices are manifold. This lets me check vertex manifoldness with essentially no overhead on `ManifoldSurfaceMesh` objects, since they can just return a buffer full of bools without explicitly checking all vertices. (Originally I thought this was unnecessary, but I was just misremembering how these classes work. This optimization provides a small, but noticeable speedup (around a few percent on the simple examples I tried)). I wasn't sure what to call this function---I settled on `SurfaceMesh::getVertexManifoldStatus()`, but I'm very open to other suggestions.


fire:
Do you require review or testing?

fire:
@MarkGillespie I was able to get this to work.

Do you know how to calculate the geodesic loop? I don't know the proper term for measuring the circumference of a human waist. This currently calculate the shortest path.

keenancrane:
@fire The MMP algorithm can’t compute geodesic loops. You can get exact geodesic loops using the FlipOut algorithm: https://nmwsharp.com/research/flip-geodesics/ Note that this implementation doesn’t handle nonmanifold meshes (though the algorithm can be generalized to this case to some degree).

PR #110, opened 2022-03-21
typo in error messages

hoskillua:
it is the same one as before in "faces mush be triangular", I'm sorry that I didn't use *find & replace* on it across all files as I didn't think it was the same message in all these files.


nmwsharp:
ha, thank you for spotting and fixing this! This is probably my fault for copy-pasting the same misspelled error message all throughout the codebase.

PR #117, opened 2022-04-20
Fixing more 404 links in docs and removing an empty page

hoskillua:
In this commit, I have gone through all pages in docs and fixed the broken links. I might have missed something, though.

Also,  here are 2 more links that I did not know what to do with:
In the [index page](https://geometry-central.net/), Yousuf Soliman's link is broken but I can't find a good alternative.
The [indexing page in utilities](https://geometry-central.net/utilities/indexing/) is not referenced in the navigation. But I don't know if that's intentional.

### DIFF
--- a/docs/docs/numerical/linear_solvers.md
+++ b/docs/docs/numerical/linear_solvers.md
@@ -10,7 +10,7 @@ These solvers provide a simple interface for solving sparse linear $Ax = b$.
 
 A key feature is that these solvers abstract over the underlying numerical library. In their most basic form, [Eigen's sparse solvers](https://eigen.tuxfamily.org/dox/group__TopicSparseSystems.html) will be used, and are always available. However, if present, the more-powerful [Suitesprase solvers](http://faculty.cse.tamu.edu/davis/suitesparse.html) will be used intead. See the [dependencies section](../../build/dependencies/#suitesparse) for instruction to build with Suitesparse support.
 
-As always, be sure to compile [with optimizations](../../build/build/#compiling-with-optimizations) for meaningful performance. In particular, Eigen's built-in solvers will be very slow in debug mode (though the Eigen QR solver is always slow).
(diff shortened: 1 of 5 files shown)


hoskillua:
Note, I am sorry that I made another PR. I did not notice them in the first one 😅.

MarkGillespie:
Thanks!

The indexing page should probably be in the navigation :) It would make sense to me to put it in the `Surface Mesh` section (and move the file accordingly).

Also, the "compile with optimizations" link in `linear_solvers.md` doesn't actually seem to describe how to compile with optimizations. It might be better to just say something like "be sure to compile with optimizations (i.e. set `CMAKE_BUILD_TYPE` to `Release`)..."

Would you be able to incorporate these changes into the PR? It would be great if you could, but if not I'm happy to accept the PR as is and move things around later

hoskillua:
@MarkGillespie Thanks for the feedback.
I have made the changes you mentioned. Also, it seems that some parts of the indexing doc are not formatted correctly, so I tried to fix that in this update as well. Please let me know if anything else can be improved :D.

MarkGillespie:
Thanks! A few tiny notes:
* in `indexing.md` in the code snippet tick marks appeared around `<bool>` in the first line. Could you delete those?
* in `indexing.md` in the "Uncached index arrays" section, there's a missing space between "and" and "_H_" (and "_C_", etc.), so they don't get formatted in italics correctly. Could you add in spaces after "and" in all these descriptions?
* in `Linear Solvers`, could you put backticks (\`) around `CMAKE_BUILD_TYPE` and `Release` so they get formatted like code?

hoskillua:
@MarkGillespie my bad 😅. Thanks for the tips and please tell me if I should edit anything else.

MarkGillespie:
Everything looks great! Thanks for making all those changes :)

PR #118, opened 2022-05-05
Cmake fixes

dyollb:
Thie PR adds the following
- install target installs the library and header files to `CMAKE_INSTALL_PREFIX`
- if user requests `BUILD_SHARED_LIBS` for a `MSVC` build symbols are automatically exported


nmwsharp:
Looks good to me, thank you for contributing!

PR #123, opened 2022-08-16
Poisson disk sampling

nzfeng:
I've added Poisson disk sampling for ManifoldSurfaceMeshes. Points are sampled so that all points are at some minimum 3d distance away from each other. 

Users can also specify SurfacePoints around which they would like to avoid samples, and specify a "radius of avoidance" describing how much samples should avoid such points. 

There's no particular acceleration, although the method takes about 2 seconds for the kitten mesh below with ~15,000 vertices (faster for less dense samplings.)

poisson_disk_sample


src/surface/poisson_disk_sampler.cpp
```diff
@@ -0,0 +1,270 @@
...
+      for (int offsetZ = -1; offsetZ <= 1; offsetZ++) {
+
+        SpatialKey bucketKey = queryKey;
+        bucketKey[0] += offsetX;
+        bucketKey[1] += offsetY;
+        bucketKey[2] += offsetZ;
+
+        auto bucketIter = spatialBuckets.find(bucketKey);
+        if (bucketIter != spatialBuckets.end()) {
+          if (bucketIter->second) {
```
nmwsharp, line 180: Is there a reason that this just checks whether the bucket is occupied, rather than storing the actual points in the buckets and checking the distance to those points?

I think that simply checking the buckets for non-emptyness could potentially lead to axis-aligned sampling artifacts, right? Because the spherical neighborhood of points within distance `r` does not exactly overlap with the cubic buckets. 

However we could instead choose the bucket size such that the 3x3x3 grid of buckets contains the sphere, then store the actual points in the buckets, and check the distance to the point inside of a any nonempty buckets before returning `false`.

Although I can imagine that any sampling bias that this strategy introduces is not a visible problem in practice.

nmwsharp:
Looks good to me! I didn't carefully check the logic but everything seems sane. Thank you for contributing!! 

You mention one more pending change to the docs, let me know when it's all done on your end and we can merge. (Also, I think it's fine to use an image rendered with a slightly-old version of the code for the docs, if that is the only change you are making. :slightly_smiling_face:)

src/surface/poisson_disk_sampler.cpp
nzfeng: Yes, you're correct. In my haste to save an extra computation, I failed to recognize that (two points in adjacent buckets are guaranteed to be within r of each other) does not imply (two points that are not in adjacent buckets are guaranteed to be greater than r away from each other.)

I've followed your suggestion, and set the side length of the buckets so that each bucket contains at most one point, and store the actual points in the buckets.

nzfeng:
All done on my end. I added another option where the user can toggle between (1) avoiding samples in a ball in 3D space around each avoidance point; or (2) avoid samples in a (2D) ball on the surface itself. Users might prefer one or the other for whatever (aesthetic) reason.

nmwsharp:
awesome, thank you!!

PR #125, opened 2022-09-13
Barycentric vectors

nzfeng:
Implemented barycentric vectors via a `BarycentricVector` class, and added documentation for them. 

I also added a minor function, `Vector2::sum()` because I wanted an easier way to sum vector components of a `Vector2` in a few functions of `BarycentricVector`.

I debugged this implementation by using them in a current project of mine; everything at least works there. I did a few simple tests for all the arithmetic and other member functions.


nmwsharp:
I only checked this briefly, but it looks good to me, and we've discussed this one on slack, so I'm assuming it's good to go!

Thank you again for contributing upstream!

PR #126, opened 2022-09-15
Crouzeix raviart Laplacians and mass matrix

nzfeng:
I've added three functions in `intrinsic_geometry_interface.cpp`, implementing a `crouzeixRaviartLaplacian`, `crouzeixRaviartMassMatrix`, and `crouzeixRaviartConnectionLaplacian`. These operators/matrix are analogous to the vertex-based cotan Laplacian and Galerkin mass matrix, but discretized using the _Crouzeix-Raviart_ (C-R) basis functions instead. The C-R basis elements are essentially Lagrange elements but based at edge midpoints: They are piecewise linear, have a value at 1 at its associated edge midpoint, and is 0 at all other adjacent edge midpoints.

In graphics/geometry processing, C-R elements have been used, for example, to [discretize bending energies](https://cims.nyu.edu/gcl/papers/wardetzky2007dqb.pdf), do [Hodge decomposition](http://www.polthier.info/articles/hodge/hodge_final.pdf), and [discretize vector Dirichlet energy](https://odedstein.com/projects/a-simple-discretization/a-simple-discretization.pdf). They can also be useful vs. the vertex Laplacians in the sense of having more DOFs, because there are typically more edges than vertices in a mesh, and there are only two faces in the support of each C-R basis function. I've used the C-R Laplacian for a couple projects now, so I figured it would be useful to add to geometry-central :)

The C-R connection Laplacian is implemented using complex numbers, in the same vein as `vertexConnectionLaplacian`.

I coded up a [demo project](https://github.com/nzfeng/crouzeix-raviart-demo) to demonstrate the C-R Laplacians in action: The first picture solves a (scalar) Poisson problem diffusing some point sources, the latter two pictures show two views of the same bunny mesh after applying the Vector Heat Method to parallel-transport an initial vector, shown in magenta. I didn't have a good method on hand for displaying vectors in edge tangent spaces, so the vector field visualization is kind of hacked up: adding a `SurfaceMesh::addEdgeIntrinsicVectorQuantity()` function to Polyscope is on my to-do list!

This implementation assumes a triangle mesh; no quad Crouzeix-Raviart yet :)

poisson-problem

parallel-transport_front
parallel-transport_back

### DIFF
--- a/docs/docs/surface/geometry/quantities.md
+++ b/docs/docs/surface/geometry/quantities.md
@@ -477,9 +477,9 @@ All operators are indexed over mesh elements according to the natural iteration 
 
 ??? func "Crouzeix-Raviart Laplacian"
     
-    ##### Crouzeix-Raviart laplacian
+    ##### Crouzeix-Raviart Laplacian
 
-    The discrete Laplace operator, discretized via the piecewise linear *Crouzeix-Raviart* elements on edges.
+    The discrete Laplace operator, discretized via the piecewise linear *Crouzeix-Raviart* basis functions associated with edge midpoints.
 
     A $|E| \times |E|$ real matrix. Always symmetric and positive semi-definite. This is the _weak_ Laplace operator, if we use it to evaluate $\mathsf{y} \leftarrow \mathsf{L} \mathsf{x}$, $\mathsf{x}$ should hold _pointwise_ quantities at edge midpoints, and the result $\mathsf{y}$ will contain _integrated_ values of the result in the neighborhood of each edge midpoint. If used to solve a Poisson problem, a mass matrix is likely necessary on the right hand side.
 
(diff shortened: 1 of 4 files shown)


nmwsharp:
looks awesome, thank you for sending this upstream!

FWIW, your blurb above about the usefulness of the CR operators is quite nice. If you'd like, I think it would be great to add it to the docs page. Including the blurb may also help non-expert users distinguish the CR Laplacian vs. the cotan Laplacian. It might be nice to start a paragraph with "The cotan-Laplacian (`getCotanLaplacian()`) is the more-common Laplace matrix in geometry problems. However, the alternative CR Laplacian has many benefits such as..." or something like that.

I'll merge this PR for now since it's good to go. But if you'd like to make another PR with that docs stuff ^^^ I'd be happy to merge it too :)

PR #127, opened 2022-09-19
Update docs for Crouzeix-Raviart operators

nzfeng:
Added a small blurb for Crouzeix-Raviart operators in the docs!


nmwsharp:
awesome, thank you!

(I'm rebuilding the public docs right now, so if you notice anything wrong/missing on the public site after 10min from now, feel free to ping me)

PR #137, opened 2022-12-10
Remeshing

MarkGillespie:
Add remeshing functions


keenancrane:
Rad! I only skimmed the changes (and read the dics) but this looks great. Note that the build seems to be failing on Mac.

lfyMeta:
Thanks, this looks useful for me as well. From a quick look, this doesn't change any logic in how edge splits/collapses currently work internally, just drives them a particular way according to the remesh code?

This is because I use data on corners (so, data on halfedges) and want to be sure my approach to fixing up halfedge data on edge collapse still works (https://github.com/nmwsharp/geometry-central/pull/135)

MarkGillespie:
Yeah, this just uses the existing split/collapse code to try to improve triangle quality

PR #145, opened 2023-03-15
Triangle mesh (self-)intersection

keenancrane:
This PR adds basic routines to detect and extract intersections between triangle meshes.  The two main routines are `surface::intersect` and `surface::selfIntersect` which extract the line segments along which two meshes intersect, or a mesh intersects itself, respectively.  See docs for more detailed information.  It also adds a utility subroutine `triTriIntersection` for intersecting a single pair of triangles.

The triangle-triangle implementation uses Tomas Moller's 1997 triangle intersection routine, provided [here](https://github.com/erich666/jgt-code/blob/master/Volume_08/Number_1/Shen2003/tri_tri_test/include/Moller97.c).  Currently this code is wrapped and used as-is, though it could be refactored to use native Geometry Central types.

**Note:** the current implementation of `intersect` and `selfIntersect` naïvely check all pairs of triangles, rather than doing fancy hierarchical testing, which will of course be slow for large models.


MarkGillespie (commented):
Looks good to me! I just have a few nitpicks:
- we should probably mention in the docs that we're using Moller's intersection code.
- most geometry-central code is formatted with `clang-format`, so I ran the formatter on this code too. Does it still look okay to you?
- it would be more idiomatic for `intersections` to take in `EmbeddedGeometryInterface`s rather than `VertexPositionGeometry`s. `EmbeddedGeometryInterface` guarantees that such a geometry provides vertex positions, while `VertexPositionGeometry` is a geometry determined by its vertex positions. In this case, the distinction is a bit silly since `VertexPositionGeometry` is the only kind of geometry which even provides vertex positions. But the analogous distinction between `EdgeLengthGeometry` and `IntrinsicGeometryInterface` (where the former geometry is determined by its edge lengths, while the latter only guarantees that it can compute edge lengths somehow) is pretty important, and it's nice to remain parallel. And I think Nick had dreams of one day extending to other kinds of geometry which might provide vertex positions without being determined by vertex positions.

keenancrane:
>we should probably mention in the docs that we're using Moller's intersection code.

Sounds good to me.

> most geometry-central code is formatted with clang-format, so I ran the formatter on this code too. Does it still look okay to you?

Yeah, totally fine.

> it would be more idiomatic for intersections to take in EmbeddedGeometryInterfaces rather than VertexPositionGeometrys.

I tried this at first, but got some complaints about it being pure virtual or something.  Do you know how to do this?

MarkGillespie:
I changed to `EmbeddedGeometryInterface` and it still builds fine for me, but I haven't tried running it on anything. Does everything still work on your computer?

nmwsharp:
Great. I just rebuilt the docs, they should be updated on http://geometry-central.net/ as soon as it percolates through the CDNs.

lfyMeta:
This looks very useful! How hard would it be to extract the intersection curves? I've never done anything with this before but I guess the tri/tri intersections must be line segments, so we can plot out the intersection line segments per triangle, merge coincident points and we'd basically have a linked list type structure of the intersection curves?

Issue #153, opened 2023-04-21
Off by one error in boundary first flattening


nervoussystem:
The boundary appears to be shifted by one index in computeBoundaryPositions making the results slightly twisted. I believe this is because it's getting the source of an Edge in the boundary rather than a Halfedge. The edges on the boundary are all oriented opposite of the half edges since they are created last. This results in a shift by one since it's starts at the head of the halfedge rather than the tail.

The last loop in the computeBoundaryPositions function could be something like this instead

```
for (BoundaryLoop b : mesh.boundaryLoops()) {
    for (Halfedge he : b.adjacentHalfedges()) {
      size_t iV = bIdx[he.vertex()];

      bdyPositions[0](iV) = x;
      bdyPositions[1](iV) = y;

      x += roundedLength(iV) * T(0, iV);
      y += roundedLength(iV) * T(1, iV);
    }
  }
```

nmwsharp:
You're absolutely right! Thanks very much for sending this up. Should be fixed in #155

PR #155, opened 2023-05-05
fix boundary indexing bug in BFF implementation

nmwsharp:


nmwsharp:
Credit to @nervoussystem in #153

PR #157, opened 2023-06-02
Minor CMake Update

qnzhou:
## Summary:
To make geometry-central more friendly as a dependency, the following cmake changes are added:
* Avoid building `nanort`, `nanoflann` and `happly` if the correspoinding cmake targets already exist.
* Move `nanoflann.h` out of the `nanoflann` directory as this is case in the [original nanoflann repo](https://github.com/jlblancoc/nanoflann). This makes geometry-central compatible with repos which interface with nanoflann directly.


nmwsharp:
Thanks for submitting! These all look good to me.

One minor request: I think the `nanoflann/nanoflann.hpp` --> `nanoflann.hpp` change could be a small breaking change for projects that included the file from the old location. It's silly, but does it sound reasonable to add a symlink at the old location just so there isn't a needless breaking change?

qnzhou:
> does it sound reasonable to add a symlink at the old location just so there isn't a needless breaking change?

Yes, I have updated the cmake to make a symlink so `#include <nanoflann/nanoflann.hpp>` will still work. :)

nmwsharp:
thank you!! merged

PR #164, opened 2023-10-17
Optimize barycentric coordinate computation

Ameobea:
In my application which traces thousands of geodesic paths, I found that the main bottleneck was solving the system of equations to convert cartesian -> barycentric coordinates.

I found a specialized solution for performing the conversion, and swapping it in produces the same results as the Eigen solver while doing so very much faster.

I totally understand if you're not interested in taking this change since it pulls code from a third party source and might have different behavior in case of weird things like degenerate/invalid triangles, NaNs, etc. but I thought I'd put it up to see if you're interested.

### DIFF
--- a/include/geometrycentral/surface/barycentric_coordinate_helpers.h
+++ b/include/geometrycentral/surface/barycentric_coordinate_helpers.h
@@ -23,7 +23,6 @@ double displacementLength2(Vector3 displacement, Vector3 triangleLengths);
 double displacementLength(Vector3 displacement, Vector3 triangleLengths);
 
 // Convert between cartesian and barycentric coordinates
-Vector3 cartesianVectorToBarycentric(const std::array<Vector2, 3>& vertCoords, Vector2 faceVec);
 Vector2 barycentricDisplacementToCartesian(const std::array<Vector2, 3>& vertCoords, Vector3 baryVec);
 
 // Normalize to sum to 1 (does nothing else)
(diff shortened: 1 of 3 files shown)


include/geometrycentral/surface/barycentric_coordinate_helpers.h
```diff
@@ -23,7 +23,6 @@ double displacementLength2(Vector3 displacement, Vector3 triangleLengths);
 double displacementLength(Vector3 displacement, Vector3 triangleLengths);
 
 // Convert between cartesian and barycentric coordinates
-Vector3 cartesianVectorToBarycentric(const std::array<Vector2, 3>& vertCoords, Vector2 faceVec);
```
nmwsharp, line 26: why delete this?

nmwsharp:
Hi! Thanks for opening this.

I'm interested in merging this, the fast method seems quite nice.

I am a liiiiitle wary because the existing colPivQR approach was very extensively numerically tested as part of our intrinsic triangulations work. The explicit version here could be more stable or less stable, but either way I am a bit reluctant to merge it in without repeating those tests (and I don't have time to repeat those tests).

What if instead we could just merge it as an alternative option, for now? As a new `cartesianVectorToBarycentric_fast()` or something, alongside the old version? That way folks can have it as an option. We could even add a flag to the `traceGeodesic()` function to use it instead.

Yvan-iComat:
@nmwsharp could you merge this PR please ? Thanks !

nmwsharp:
Hi @Yvan-iComat 

For now my reasoning in the comment above still applies---this is a numerically-important subroutine, so I'm reluctant to change the default method to something that I suspect is less numerically-stable even if it is faster. I would be open to merging this as an optional alternate method, however.

Can you say what you want it for? Is it for the geodesic tracing or something else?

PR #168, opened 2023-12-21
Dual mesh and dual geometry.

filthynobleman:
Added functionalities for computing dual meshes and geometries. The functions also handle boundaries and leave to the user the option to preserve either the bijection between primal vertices and dual faces or between primal faces and dual vertices.
Preserved bijections also preserve the indexing (i.e., primal vertex with index i is mapped to dual face with index i, and primal face with index j is mapped to dual vertex with index j).
I think the code can be optimized, but I have extensively tested it in a personal project and it produces correct results.

### DIFF
--- a/include/geometrycentral/surface/dual_mesh.h
+++ b/include/geometrycentral/surface/dual_mesh.h
@@ -1,0 +1,82 @@
+#pragma once
+
+
+#include "geometrycentral/surface/manifold_surface_mesh.h"
+#include "geometrycentral/surface/vertex_position_geometry.h"
+
+
+namespace geometrycentral
+{
+namespace surface
+{
+
+/**
+ * @brief       Builds the dual manifold mesh of the input manifold mesh.
+ * 
+ * @details     This function computes the dual manifold mesh of the given primal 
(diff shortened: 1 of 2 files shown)


patrick-laurin:
What applications would such a feature have? I'm genuinely curious!

filthynobleman:
Personally, I've used the feature for visualizing Voronoi regions onto the surface. With standard vertex coloring I got smooth transitions between regions, and with triangle coloring the result is technically incorrect. Using face coloring on the dual mesh (keeping the one-to-one correspondence between primal vertices and dual faces), I was able to visualize the partitioning correctly.

As a side note, I've also used it for some other generic visualizations of primal/dual correspondences, for batch generation of polygonal meshes, and for testing the consistency of some methods on polygonal and triangular meshes.

adsk-villeneuveb:
> What applications would such a feature have? I'm genuinely curious!
I`m late here but, 
The octahedron is the dual of the cube, both platonic solids. The dual operation is very useful to procedurally generate solids from primitives. 

Many catalans solids are the dual of some archemedean solids. 
image

PR #170, opened 2024-01-31
Updating the functionality of barycentric vectors

nzfeng:
Nothing major -- just added additional functionality to the `BarycentricVector` object:
- `rotated90()` now supports edge-type vectors as input
- a new function `rotated()` allows rotation of vectors by arbitrary angles
- added some more detailed comments in the code so I don't forget how I derived quantities :)
- documentation updated accordingly.

Let me know if you need me to add additional test functions to geometry-central -- for these few commits, I tested everything using an external program: https://github.com/nzfeng/barycentric-vector-demo

### DIFF
--- a/include/geometrycentral/surface/barycentric_vector.h
+++ b/include/geometrycentral/surface/barycentric_vector.h
@@ -78,6 +78,9 @@ struct BarycentricVector {
   BarycentricVector rotated90(IntrinsicGeometryInterface& geom) const; // return a new vector
 };
 
+// Helper function for rotated90()
+BarycentricVector faceVectorRotated90(const BarycentricVector& v, IntrinsicGeometryInterface& geom);
+
 // Returns an arbitrary face shared by two vectors, if one exists; returns Face() if none.
 Face sharedFace(const BarycentricVector& u, const BarycentricVector& v);
 // Returns the edge shared by two vectors, if one exists; return Edge() if none.
--- a/include/geometrycentral/surface/barycentric_vector.h
+++ b/include/geometrycentral/surface/barycentric_vector.h
@@ -76,10 +76,14 @@ struct BarycentricVector {
   // Rotate the vector 90 degrees CCW within the face it belongs to.
   // This requires the geometry, since the meaning of "90 degrees" depends on the geometry of the triangle.
   BarycentricVector rotated90(IntrinsicGeometryInterface& geom) const; // return a new vector
+  // Rotate the vector the specified amount of radians CCW within the face it belongs to.
+  // This again requires the geometry.
+  BarycentricVector rotated(IntrinsicGeometryInterface& geom, double angle) const;
 };
 
-// Helper function for rotated90()
+// Helper functions for rotations
 BarycentricVector faceVectorRotated90(const BarycentricVector& v, IntrinsicGeometryInterface& geom);
+BarycentricVector faceVectorRotated(const BarycentricVector& w, IntrinsicGeometryInterface& geom, double angle);
 
 // Returns an arbitrary face shared by two vectors, if one exists; returns Face() if none.
 Face sharedFace(const BarycentricVector& u, const BarycentricVector& v);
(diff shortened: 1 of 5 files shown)


include/geometrycentral/surface/barycentric_vector.h
```diff
@@ -76,8 +76,15 @@ struct BarycentricVector {
   // Rotate the vector 90 degrees CCW within the face it belongs to.
   // This requires the geometry, since the meaning of "90 degrees" depends on the geometry of the triangle.
   BarycentricVector rotated90(IntrinsicGeometryInterface& geom) const; // return a new vector
+  // Rotate the vector the specified amount of radians CCW within the face it belongs to.
+  // This again requires the geometry.
+  BarycentricVector rotated(IntrinsicGeometryInterface& geom, double angle) const;
```
nmwsharp, line 81: Is there a reason for 'rotated' rather than 'rotate' here? In some other libraries, present-tense vs. past-tense is used to indicated modification in-place vs returning  a new vector, although I think geometry-central does not use that convention and just always returns a new vector.

In the `Vector2` class we use 'rotate' https://github.com/nmwsharp/geometry-central/blob/master/include/geometrycentral/utilities/vector2.h#L62, so I would prefer to stick with that tense if possible.

nmwsharp: Oh, I guess maybe `rotated90` already existed, so you're following that convention?

That probably have been `rotate90` 😅

include/geometrycentral/surface/barycentric_vector.ipp
```diff
@@ -195,35 +195,58 @@ inline BarycentricVector BarycentricVector::rotated90(IntrinsicGeometryInterface
...
 }
 
+inline BarycentricVector BarycentricVector::rotated(IntrinsicGeometryInterface& geom, double angle) const {
+
+  switch (type) {
+  case BarycentricVectorType::Face: {
+    return faceVectorRotated(*this, geom, angle);
+    break;
+  }
+  case BarycentricVectorType::Edge: {
```
nmwsharp, line 225: Do I read correctly that this takes in a `BarycentricVectorType::Edge`, but returns a `BarycentricVectorType::Face`? To me that would be confusing behavior, could we send it back to the edge before returning?

nmwsharp: (same for the other switch statement above)

nmwsharp: Ah this is well-documented though. I guess my question is _why_ do we want to return a vector on the adjacent face rather than keeping it on the edge?

nmwsharp:
This looks great! Docs are good too. I left two small comments above, but neither are big issues, just small questions.

include/geometrycentral/surface/barycentric_vector.h
nzfeng: I was under the impression that past tense is used to indicate returning a new vector, while present tense indicates in-place modification, for example [Eigen's `normalize()` vs. `normalized()` convention](https://eigen.tuxfamily.org/dox/group__QuickRefPage.html):
image
Admittedly my decision was solely based this convention, and maybe Eigen is the exception! I'm fine with changing the tense to match the rest of geometry-central, though.

include/geometrycentral/surface/barycentric_vector.ipp
nzfeng: Good question -- I admit it feels weird. I guess there's a subtle technicality: This function is rotating a tangent vector, located in the tangent space of the edge, and this vector happens to be represented via barycentric coordinates on the edge. But edge-type vectors are constrained to point along the edge, since the space of edge-type vectors is only one-dimensional: there are two (dependent) barycentric coordinates, one corresponding to each vertex of the edge. So to do this rotation, we interpret the vector as living in its 2d tangent space, and re-interpret the rotated vector as a barycentric vector living in one of the adjacent faces. I'm open to ideas on how to make this more intuitive!

nmwsharp:
@nzfeng just realized this was never merged, is it good-to-go on your end? if so we can hit the merge button

nzfeng:
Yes, it's good to go on my end!

nmwsharp:
Awesome! thank you!!

nmwsharp:
Docs are already updated here: https://geometry-central.net/surface/utilities/barycentric_vector/

PR #171, opened 2024-02-16
Add Bazel support for geometry-central

phaedon:
This PR adds support for Bazel, as an alternative to CMake (https://bazel.build/about/why)

To build:
`bazel build ...`

To run all unit tests:
`bazel test ...`

Next steps (after this is merged):

1. Please cut a new release on https://github.com/nmwsharp/geometry-central/releases
2. Then I can add this repo to https://registry.bazel.build/
3. I will add documentation to this repo's `building.md` to document the one-liner required to pull this into any other Bazel-based project (which will look something like `bazel_dep(name = "geometry_central", version = "0.2.0")`)


phaedon:
Verified that this builds and runs successfully on at least Ubuntu (x86_64) and Mac OS (ARM).

nmwsharp:
Thanks for submitting this! I'm totally unfamiliar with Bazel, but it seems to be fairly widely used so it'd be nice to support.

One question: what is the maintenance requirement going forward to ensure this library continues to build well on Bazel? I'm quite wary of leaving behind stale/broken versions of the library on the long tail of various package repositories / build systems. Does it just mean updating the `BUILD.bazel` as files are added? Would it make sense to also add a bazel build to geometry-central's CI matrix?

nmwsharp:
Also what is the `.bazelrc` doing? Do we need to add an additional file to the project root?

phaedon:
Great suggestion, I added a `bazel.yml` file to run the build and test stages on each PR.

The `.bazelrc` specifies a couple of basic but important settings for build:
- C++20 version
- "bzlmod" which reads the external dependencies (eigen and googletest) from the MODULE.bazel file
as well as test output only on errors. I tried to add documentation to that file.

phaedon:
Just a reminder that this is ready to go in.

phaedon:
Hi again, I can actually add this project to the Bazel Central Registry (BCR) without this MR. I'm happy to close it out and at this point (given the evolution of Bazel) it's a more sensible approach. However, I need you to post another release archive; the last one was in 2021 so it's quite out of date. Could you please do this soon?

phaedon:
Closing this out, in favour of adding the Bazel rules for this repo directly to the Bazel Central Registry. Thanks!

PR #178, opened 2024-06-07
Fix SuiteSparse call for underdetermined systems

zoemarschner:
Currently, solving underdetermined systems using SuiteSparse by calling `solve` does not work. For example,

```
#include "geometrycentral/numerical/linear_solvers.h"

Eigen::Matrix<double, 2, 3> A;
A << 1, 0, 1, 0, 1, 0 ;
Vector<double> b(2);
b << 1, 2;
SparseMatrix<double> sparseA = A.sparseView();

Vector<double> sol = solve(sparseA, b);

std::cout << sol << std::endl;
```

will print out `1 1 1`, but the correct answer is `0.5 2 0.5`. This is because SuiteSparse's QR decomposition includes a permutation matrix (i.e., `E` such that `QR=AE`), but geometry central does not properly account for this in the call to `SuiteSparseQR_solve`. (Thank you @MarkGillespie for helping to find the bug!)

### DIFF
--- a/src/numerical/qr_solvers.cpp
+++ b/src/numerical/qr_solvers.cpp
@@ -151,11 +151,15 @@ void Solver<T>::solve(Vector<T>& x, const Vector<T>& rhs) {
   // Solve
   // outVec = SuiteSparseQR<typename Solver<T>::SOLVER_ENTRYTYPE>(cMat, inVec, internals->context);
   // Note that the solve strategy is different for underdetermined systems
+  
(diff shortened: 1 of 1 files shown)


nmwsharp:
Looks great, thank you so much for this! I haven't tested this code, but the problem makes sense and the fix looks reasonable so I'm going to merge it.

Also, I love that this removes a `// TODO what is this E doing here?` comment 😂

PR #181, opened 2024-07-17
feat: Add convex embedding algorithm

keenancrane:
embed

This PR adds an algorithm to geometry-central for embedding an intrinsic triangulation with a convex metric into three-dimensional space.  It implements the strategy of [Bobenko and Izmestiev](https://arxiv.org/abs/math/0609447) and draws inspiration from the [Java implementation](https://gitlab.discretization.de/sechel/f1software/-/tree/master/src/alexandrov) by Stefan Sechelmann (though deviates a fair bit from the latter).  This method fits nicely into Geometry Central since it builds on its existing support for intrinsic triangulations, which are a central part of Bobenko and Izmestiev's algorithm.

The main additions/changes are:

* `surface/embed_convex.*` — implements the convex embedding algorithm
* `utility/elementary_geometry.*` — adds some basic methods for tetrahedra
* `docs/surface/algorithms/embedding.md` — documentation for embedding

See the documentation for more details.  Note that documentation is named `embedding.md` rather than `convex_embedding.md` in case we want to support other embedding methods in the future (like [this one](https://cseweb.ucsd.edu/~alchern/projects/ShapeFromMetric/) for instance).

There should be no interactions or side effects with other geometry-central modules.


docs/docs/surface/algorithms/embedding.md
```diff
@@ -0,0 +1,164 @@
...
+
+```cpp
+EmbedConvexOptions options;
+options.tolerance = 1e-2; // use a looser tolerance than default
+EmbedConvexResult result = embedConvex( *mesh, *uvs, options );
+```
+
+??? func "`#!cpp EmbedConvexResult embedConvex(ManifoldSurfaceMesh& mesh, EdgeData<double>& edgeLengths, EmbedConvexOptions options = defaultEmbedConvexOptions);`"
+
+    Embed a metric described by a set of edge lengths.  Options are passed as a [RemeshOptions](#options) object.  If embedding fails, the flag `EmbedConvexResult::success` will be set to `false`.
```
nmwsharp, line 54: Might be useful to indicate in these docs whether the input mesh/edge lengths are changed, or if they are left unchanged. (Judging by the other docs I assume they are unchanged, and the `EmbedConvexResult` holds the solution.)

Another option could be to mark the inputs as `const`,  but it might not be that easy as some of the other geometry-central container methods are not marked const even though you'd think they would be, because data structures get automatically/lazily resized/initialized under the hood.

nmwsharp:
Awesome!! I just scrolled the code and left one (unimportant) comment. Looks great and good-to-merge to me. @MarkGillespie can chime in if he has any comments, otherwise I'll merge soon and re-deploy the docs.

(FYI I also pushed a commit after running a formatter on the code according to the `.clang-format` settings used with the rest of geometry-central, just to stay consistent)

docs/docs/surface/algorithms/embedding.md
keenancrane: Yeah, I had actually written the arguments as const at first but hit some compiler errors and didn't want to do meticulous constifying. Adding a note to the header/docs is a good idea.

keenancrane:
Thanks Nick!

MarkGillespie:
It might be nice to add an option that will suppress all of the printed output. But other than that, it looks good to me

keenancrane:
Thanks @MarkGillespie!

I added a `verbose` flag to the options; most output is now suppressed by default.

I also noted that input data is not changed, as suggested by @nmwsharp.

Barring any further observations, we should be good to go.

MarkGillespie:
Looks good to me!

keenancrane:
Great.  Updated with `master`, and merging PR.  Thanks all!

keenancrane:
@nmwsharp Not sure what has to happen to update docs (or if this is something only you can trigger).

MarkGillespie:
The geometry-central docs are managed through this other repository: https://github.com/nmwsharp/geometry-central-docs. It has instructions for updating them, but Nick probably needs to give you access to the repo first

nmwsharp:
Awesome!!

Yep, as Mark says we use the the `geometry-central-docs` repo to host the actual docs (it's just a simple mirror of this repo which also holds the compiled HTML and has github pages enabled for hosting).

I just invited @keenancrane and also @GeometryCollective to that repo. There's instructions in the readme to update the public facing page.

nmwsharp:
I just ran the docs update, it's live here: https://geometry-central.net/surface/algorithms/embedding/

PR #182, opened 2024-08-04
make vector2 rotation direction explicit, deprecate CW default

nmwsharp:


keenancrane:
LGTM. The move to explicit CW and CCW is probably the right one. I guess people who are already using rotate() might be a bit surprised to find it missing from the docs, but when they go searching in the headers will find the right message (DEPRECATED). Thanks @nmwsharp!

PR #185, opened 2024-08-09
numerical robustness fixes for quadric simplification

nmwsharp:


nmwsharp:
@MarkGillespie not urgent, but if you want to take a look the important change here is

```
Eigen::Vector3d Quadric::optimalPoint() { return -A.inverse() * b; }
```
to
```
Eigen::Vector3d Quadric::optimalPoint() { return -A.colPivHouseholderQr().solve(b); }
```

for numerical stability. It resolves some NaNs I was seeing even on nice inputs. Seem reasonable?

Issue #183, opened 2024-08-08
Make `meanEdgeLength` public in PoissonDiskSampler


dyollb:
The parameter `rCoeff` in PoissonDiskSampler is used to compute the minimum distance between points
```cpp
minDist = rCoeff * meanEdgeLength;
```

The value `meanEdgeLength` is computed by `PoissonDiskSampler` and stored as member. If I want to sample a surface with a known `minDist` I need to compute the `meanEdgeLength` myself in order to specify `rCoeff` as `minDist / meanEdgeLength`.
I would propose to add a getter to get the mean edge length, e.g.

```
class PoissonDiskSampler {
...
double getMeanEdgeLength() const { return meanEdgeLength; }
...
};
```

dyollb:
fixed by #186

PR #186, opened 2024-08-16
fix bug in PoissonDiskSampler, add test

### LINKED ISSUE #184: Weird behavior in PoissonDiskSampler
I am trying to use the PoissonDiskSampler to sample a triangle mesh with minimum sample distance `distance`:

```cpp
    gc::surface::SimplePolygonMesh simpleMesh;
    for (const auto & t : qmesh.get_tris_buffer())
    {
        simpleMesh.polygons.push_back({t.inds[0], t.inds[1], t.inds[2]});
    }
    for (const auto & v : qmesh.get_verts_buffer())
    {
        simpleMesh.vertexCoordinates.push_back({v.x, v.y, v.z});
    }

    std::unique_ptr<gc::surface::ManifoldSurfaceMesh> mesh;
    std::unique_ptr<gc::surface::VertexPositionGeometry> geometry;
    // throws exception if input is non-manifold/invalid
    std::tie(mesh, geometry) = gc::surface::makeManifoldSurfaceMeshAndGeometry(simpleMesh.polygons, simpleMesh.vertexCoordinates);

    geometry->requireEdgeLengths();
    double meanEdgeLength = 0.;
    for (auto e : mesh->edges()) {
      meanEdgeLength += geometry->edgeLengths[e];
    }
    meanEdgeLength /= mesh->nEdges();
    geometry->unrequireEdgeLengths();

    gc::surface::PoissonDiskSampler sampler(*mesh, *geometry);
    auto samples = sampler.sample(distance / meanEdgeLength);

    std::vector<Vec3d> points;
    points.reserve(samples.size());
    for (auto& s: samples)
    {
        auto p = s.interpolate(geometry->vertexPositions);
        points.push_back({p.x, p.y, p.z});
    }
```

If I sample a (approximate) cube with edge length 40 (two triangles per cube face), I get following behavior:

| distance=4  | distance=2 | distance=1 |
| :---: | :---: | :---: |
| samples: 3 | samples: 22 | samples: 6559 |
| image  | image  | image |

Why is this happening? Am I doing something wrong?

dyollb:
Fixed #184 

Sampling only behaved reasonable when the mean edge length was close to 1, because the "rMinDist" was treated as if it where the squared min. distance (sqrt was taken).

This PR fixes this issue and adds a test.
- a 1x1 square disk is sampled with distance=0.1
- a 100x100 square disk is sampled with distance=10 (i.e., 100*0.1)

In both cases we expect a similar number of points (seed cannot be set from API).

Without the fix, I get approx. 650 (width=1) vs 10 (width=100) points.

### DIFF
--- a/docs/docs/surface/algorithms/surface_sampling.md
+++ b/docs/docs/surface/algorithms/surface_sampling.md
@@ -46,7 +46,7 @@ std::vector<SurfacePoint> samples = poissonSampler.sample(); // sample using def
 
 // Sample with some different parameters.
 PoissonDiskOptions sampleOptions;
-sampleOptions.r = 2.;
+sampleOptions.minDist = 2.;
 std::vector<SurfacePoint> newSamples = poissonSampler.sample(sampleOptions);
 ```
 ## Helper Types
@@ -56,10 +56,10 @@ Options are passed in to `options` via a `PoissonDiskOptions` object.
 
 | Field | Default value |Meaning|
 |---|---|---|
-| `#!cpp double r`| `1` | The minimum distance between samples, expressed in world-space units. |
+| `#!cpp double minDist`| `1` | The minimum distance `r` between samples, expressed in world-space units. |
 | `#!cpp int kCandidates`| `30` | The number of candidate points chosen from the (`r`,2`r`)-annulus around each sample. |
 | `#!cpp std::vector<SurfacePoint> pointsToAvoid`| `std::vector<SurfacePoint>()` | Points which samples should avoid. |
-| `#!cpp double rAvoidance`| `1` | The radius of avoidance around each point to avoid, expressed in world-space units. |
+| `#!cpp double minDistAvoidance`| `1` | The radius of avoidance around each point to avoid, expressed in world-space units. |
 | `#!cpp bool use3DAvoidance`| `true` | If true, the radius of avoidance will specify a solid ball in 3D space around which samples are avoided. Otherwise, samples are avoided within a geodesic ball on the surface. |
 
-Using the `use3DAvoidance` option, the radius of avoidance `rAvoidance` can specify either the radius in 3D space, or in terms of distance along the surface. The former will produce a radius of avoidance that will appear perfectly round, and is likely more visually pleasing, but for very large radii may occlude samples from opposite sides of the mesh. The latter will restrict the radius of avoidance to only be along the surface, but such a metric ball may not appear perfectly round, especially in areas with very large changes in curvature.
+Using the `use3DAvoidance` option, the radius of avoidance `minDistAvoidance` can specify either the radius in 3D space, or in terms of distance along the surface. The former will produce a radius of avoidance that will appear perfectly round, and is likely more visually pleasing, but for very large radii may occlude samples from opposite sides of the mesh. The latter will restrict the radius of avoidance to only be along the surface, but such a metric ball may not appear perfectly round, especially in areas with very large changes in curvature.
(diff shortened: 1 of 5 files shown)


keenancrane:
LGTM. @nmwsharp Should we be worried about this being a breaking change, e.g., for algorithms or demos that assume the original behavior (even it was wrong) in setting parameters, etc.?

nzfeng:
Also looks good to me -- thank you for adding tests too!

One thing I've been meaning to do is to expose the distance parameter itself like @dyollb suggested, since in most cases it doesn't make much sense to have the minimum distance be in terms of the mesh's edge length. (I originally coded it up this way as a hacky proxy to a good "visual sampling density".) What do you all think?

dyollb:
> Also looks good to me -- thank you for adding tests too!
> 
> One thing I've been meaning to do is to expose the distance parameter itself like @dyollb suggested, since in most cases it doesn't make much sense to have the minimum distance be in terms of the mesh's edge length. (I originally coded it up this way as a hacky proxy to a good "visual sampling density".) What do you all think?

I pushed a small change to expose the mean edge length (fixes issue #183 )

nmwsharp:
Thank you for pointing this out, fixing it, and even adding a test!

> Breaking change

This is definitely a bug, and it's somewhat-okay to make breaking changes for bugfixes. However, this is indeed the kind of change that would silently cause someone's code to stop working on update, and it's best to avoid those whenever possible. 

How about we also take the opportunity to factor out these options into a little options struct, and additionally include an `absoluteRadius` option? Any callers will get a compile error and be forced to briefly look at the code and chose a new parameter, and we can also offer the more-useful world coordinates radius an option.

(if this sounds like a good idea but others don't have time, I'd be happy to make this change)

dyollb:
> How about we also take the opportunity to factor out these options into a little options struct, and additionally include an `absoluteRadius` option? Any callers will get a compile error and be forced to briefly look at the code and chose a new parameter, and we can also offer the more-useful world coordinates radius an option.
> 
> (if this sounds like a good idea but others don't have time, I'd be happy to make this change)

I added a struct for the options. Should we rename the options to something users might understand without reading the code? e.g.,

```cpp
struct PoissonDiskOptions {
  double relativeRadius = 1.0;
  double absoluteRadius = -1.0;
  int numCandidates = 30;
  std::vector<SurfacePoint> pointsToAvoid;
  int relativeAvoidanceRadius = 1;
  bool use3DAvoidanceRadius = true;
};
```

dyollb:
Unfortunately, the test is [flaky](https://github.com/nmwsharp/geometry-central/actions/runs/10466242263/job/28982786894?pr=186)
> Expected: (n1) > (600), actual: 592 vs 600

I can try to modify the test so it is not flaky, or allow the user to set a seed for the random number generator. What do you think?

dyollb:
> I can try to modify the test so it is not flaky, or allow the user to set a seed for the random number generator. What do you think?

I now set the seed in the test. Number of samples is identical for the scaled/unscaled square disk.

nzfeng:
Thanks @dyollb for adding the new feature! Agree that adding a struct for options is the way to go.

I've streamlined the struct to only use distances specified world-space units, and updated the docs.

The struct is now
```
struct PoissonDiskOptions {
  double r = 1.0;       // minimum distance between samples
  int kCandidates = 30; // number of candidate points chosen from the (r,2r)-annulus around each sample
  std::vector<SurfacePoint> pointsToAvoid;
  double rAvoidance = 1.0; // radius of avoidance
  bool use3DAvoidance = true;
};
```
I'm strongly in favor of using only absolute distances, to avoid confusion, and to let the user decide if they want to specify values in terms of edge length, etc. I've tested everything on my end, so if there's no other bugs/features we'd like to fix, I'll merge.

dyollb:
LGTM. Maybe the options could have more expressive names, e.g., 'r' -> 'rsdius' or 'sample_distance'.

nzfeng:
I've just merged. Thanks again for your help!

nmwsharp:
Thanks a bunch @dyollb and @nzfeng, this is great!! I just rebuilt the docs, they're updated to show this change https://geometry-central.net/surface/algorithms/surface_sampling/

PR #195, opened 2024-09-17
Polygon DEC

nzfeng:
Adds polygon Laplacian and other operators for general polygon meshes. 

Summary of contributions:
- implementation of polygon Laplacian and mass matrices (unlumped & lumped) from [Polygon Laplacian Made Simple](https://www.cs.jhu.edu/~misha/MyPapers/EUROG20.pdf) by Bunge et al. 2020.
- implementation of polygon Laplacian, mass matrix, gradient, divergence, connection Laplacian, and DEC operators from [Discrete Differential Operators on Polygonal Meshes](https://graphics.pixar.com/library/PolyDDG/paper.pdf) by de Goes et al 2020.
- A `PolygonMeshHeatSolver` that implements the Vector Heat Method, Unsigned Heat Method, and Signed Heat Method on polygon meshes, analogous to the `PointCloudHeatSolver` for point clouds.
- Documentation & tests.

image

The only thing I'm not quite satisfied with is the naming convention for the operators from "Polygon Laplacian Made Simple": `geometry.simplePolygonLaplacian`, `geometry.simplePolygonVertexLumpedMassMatrix`, and `geometry.simplePolygonVertexGalerkinMassMatrix`. My concerns are

- the names can be quite long;
- perhaps the word "simple" will be mistaken as meaning "less sophisticated" and therefore "less accurate";
- perhaps the operators will be mistaken as only applying to "simple polygons", which are non-self-intersecting polygons without holes, though they can indeed apply even when polygon faces are self-intersecting.

Suggestions & review are welcome!


nzfeng:
@MarkGillespie or @nmwsharp, adding this support for polygon meshes required making some changes for `EmbeddedGeometryInterface` --- mainly removing the asserts that assumed triangular faces in functions like `computeCornerAngles()` and `computeFaceAreas()`. 

I don't foresee that these changes will result in breaking behavior (especially since all tests have passed), but it might also be a good idea to have a geometry-central expert check the details in `surface/embedded_geometry_interface.cpp`!

fire:
> perhaps the word "simple" will be mistaken as meaning "less sophisticated" and therefore "less accurate";

What do you think of this argument? 

geometry.basePolygonLaplacian, geometry.basePolygonVertexLumpedMassMatrix, and geometry.basePolygonVertexGalerkinMassMatrix.

Rename `simple` to `base`.

Like the base polygons are unrestricted, while for example manifold polygon has numerous restrictions.

include/geometrycentral/surface/surface_point.ipp
nmwsharp, line 1: Why did we define == with a hidden internal `eps`!?? If I did that it was a silly mistake... This is breaking change but a good one I think.

include/geometrycentral/surface/surface_point.ipp
```diff
@@ -263,11 +263,11 @@ inline bool SurfacePoint::operator==(const SurfacePoint& other) const {
     break;
   }
   case SurfacePointType::Edge: {
-    return edge == other.edge && abs(tEdge - other.tEdge) < eps;
+    return edge == other.edge && tEdge == other.tEdge;
```
nmwsharp, line 266: can you delete the unused `eps` variable above in this function?

include/geometrycentral/surface/embedded_geometry_interface.h
```diff
@@ -84,6 +135,96 @@ class EmbeddedGeometryInterface : public ExtrinsicGeometryInterface {
...
+  // Vertex connection Laplacian
+  DependentQuantityD<Eigen::SparseMatrix<std::complex<double>>> polygonVertexConnectionLaplacianQ;
+  virtual void computePolygonVertexConnectionLaplacian();
+
+  // DEC Operators
+  std::array<Eigen::SparseMatrix<double>*, 7> polygonDECOperatorArray;
+  DependentQuantityD<std::array<Eigen::SparseMatrix<double>*, 7>> polygonDECOperatorsQ;
+  virtual void computePolygonDECOperators();
+
+  // helper functions
```
nmwsharp, line 196: These are a _lot_ of helper functions. I'm worried this class/file is getting too busy.

Could we maybe move these helpers to a separate file, and invoke them as-needed? And maybe once moved to their own file they no-longer need to be virtual?

src/surface/embedded_geometry_interface.cpp
```diff
@@ -278,11 +288,11 @@ void EmbeddedGeometryInterface::computeCornerAngles() {
     Vector3 pA = vertexPositions[he.vertex()];
     he = he.next();
     Vector3 pB = vertexPositions[he.vertex()];
-    he = he.next();
+    do {
```
nmwsharp, line 291: This function has quadratic complexity in the face degree, when it doesn't need to. It could be implemented with linear complexity by working one face at a time and maintaining a leading a trailing reference while orbiting the face.

[doesn't need to be fixed, just couldn't help but mention it :)]

nmwsharp (commented):
This is awesome!! I left a few comments, but nothing major. Looks great.

nmwsharp:
Re naming for `simple` polygon Laplacian (thanks for the ideas @fire!). Personally I think the name 'simple' is perfectly good. It matches the title of the paper it comes from. We also already use the term elsewhere in `SimplePolygonMesh`. Any confusion about the term will be quickly resolved by reading the docs.

src/surface/embedded_geometry_interface.cpp
nzfeng: Should be fixed in a most recent commit!

include/geometrycentral/surface/surface_point.ipp
nzfeng: Done!

include/geometrycentral/surface/surface_point.ipp
nzfeng: I think that might have actually been me at some point in the past... sorry! (I might have done this while debugging something else and accidentally included it in a PR.)

nzfeng:
FYI this PR also updates the `cornerAngles` issue mentioned [here](https://github.com/nmwsharp/geometry-central/issues/214#issue-3066729414).

include/geometrycentral/surface/embedded_geometry_interface.h
nzfeng: Yeah... a lot of these functions depend on `EmbeddedGeometryInterface` and quantities within, which made it hard to truly separate these functions into separate files without writing tangled code with a bunch of arguments accounting for data being passed around (commit [c94b23c](https://github.com/nmwsharp/geometry-central/pull/195/commits/c94b23c856786fdc7925d889cf6f3f2bdc636b90)). 

For now I separated what I could cleanly separate, into a new file called `polygon_mesh_helpers.cpp`, which doesn't help a ton but is at least a *tiny* bit better. Other options I can see are to (1) move some function definitions into separate `cpp` files, even if all the function declarations remain in the one file `embedded_geometry_interface.h`; (2) think long and hard about how the number of functions could be reduced... though I'm not sure how to do that, currently, without sacrificing code cleanliness :(

nzfeng:
I've squash-merged for now! 🤞

PR #201, opened 2024-11-12
Move O(n) operations to the initializer in exact_geodesics

yutanoma:
Fixed `exact_geodesics` so that you do not need to run O(n) operations whenever you query the geodesic path.
This is especially useful when you have a fixed mesh/geometry and want to query the geodesic path multiple times.

## Summary of contribution

- Moved `geom.requireCornerAngles()` and `geom.requireVertexGaussianCurvatures()` to the initializer
- Added `clear_data()` that clears the data except the ones associated with the mesh/geometry

## Usage

```c++
std::vector<SurfacePoint> p0s, p1s;
// initialize your p0s/p1s here

GeodesicAlgorithmExact mmp(mesh, geometry);

for (int i = 0; i < p0s.size(); i++) {
  auto p0 = p0s[i];
  auto p1 = p1s[i];

  mmp.propagate({p0}, GEODESIC_INF, {p1});
  std::vector<SurfacePoint> path = mmp.traceBack(p1);
  // do science with your path here

  // clear the propagated data so that you can use `mmp` for another geodesic computation
  mmp.clear_data();
}
```

### DIFF
--- a/include/geometrycentral/surface/exact_geodesics.h
+++ b/include/geometrycentral/surface/exact_geodesics.h
@@ -100,6 +100,7 @@ protected:
   bool check_stop_conditions(unsigned& index) const;
 
   void clear();
+  void clear_data();
 
   list_pointer interval_list(Edge e) { return &m_edge_interval_lists[e]; };
   const_list_pointer interval_list(Edge e) const { return &m_edge_interval_lists[e]; };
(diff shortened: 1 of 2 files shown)


nmwsharp:
Hey, looks like I forgot to merge this from ages ago -- my apologies!

Looking at it now, I have a question, it's not clear to me what `clear()` vs the new `clear_data()` do? Do you remember? I want to add some guidance for the user about when to call one vs. the other.

Is it that `clear_data()` needs to be called when propagating from a different source set? I'm a bit confused because it seems like `clear()` is already called at beginning of each `propagate()` and resets a bunch of data.

nmwsharp:
I started to clean this up to merge, but then I realized it doesn't really make sense to me. I don't see how it improves O(n) behavior for multiple paths.

I'm going to close this for now assuming it's stale code. Just let me know if I have misunderstood and we can reopen it.

PR #206, opened 2025-01-14
add cmake exported targets

aschier:
Add exported targets to make it easier to link against geometry-central.

Projects can then simply use
```cmake
find_package(geometry-central)
```
to link against it and configure the location using `geometry-central_DIR`, which will also be automatically detected if geometry-central is installed in `$CMAKE_PREFIX_PATH`.
I also added the version (from the latest tag) to the CMakeLists.txt so that projects can require a specific version.

The targets should probably be handled by `${GC_DEP_LIBS}`, but that causes a mismatch between the `eigen` and `Eigen3::Eigen` targets, and I do not want to change too much in your CMakeFiles.txt files. Possibly these dependencies could also be private in case that geometry-central links them statically.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,9 @@
 cmake_minimum_required(VERSION 3.14.0)
+include(CMakePackageConfigHelpers)
+include(GNUInstallDirs)
 
 project(geometry-central)
+set(PROJECT_VERSION 0.1.0)
 
 ### Policy settings
 cmake_policy(SET CMP0054 NEW)   # don't implicitly dereference inside if()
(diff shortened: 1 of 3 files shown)


hollowsunhc:
Looks similar to what I'm achieving here https://github.com/nmwsharp/geometry-central/pull/176

aschier:
I'd happily use your commit when it gets merged. This one is the minimalist version to get it to run.
What do the maintainers think? There is little news on the other pull request as well.

nmwsharp:
Hi! I just merged another PR addressing some of the same issues. https://github.com/nmwsharp/geometry-central/pull/238

From my understanding this should be sorted out now. Please let me know if not! Otherwise I will close this PR next time I come back to it.

aschier:
Looks good to me

PR #207, opened 2025-01-22
Compatibility with recent nanoflann versions.

jdumas:
- v1.5.0: Changed output arg from `std::pair<>` to `nanoflann::ResultItem<>`
- v1.5.0: `naonflann::SearchParams` renamed to `naonflann::SearchParameters` (because the first arg in the constructor was also removed from the API).
- v1.4.0: default template arg for `IndexType/AccessorType` was changed from `size_t` to `uint32_t`:
  https://github.com/jlblancoc/nanoflann/pull/154

I haven't updated the nanoflann version bundled with geometry-central, but I've confirmed that it compiles locally with the latest version.

### DIFF
--- a/src/utilities/knn.cpp
+++ b/src/utilities/knn.cpp
@@ -25,7 +25,7 @@ struct Vector3Adaptor {
 class NearestNeighborFinder::KNNImpl {
 public:
   // == Types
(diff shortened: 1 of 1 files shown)


phaedon:
This would be great; I had to add support for a really old version of nanoflann in order to check this library into the Bazel registry:
 
https://github.com/bazelbuild/bazel-central-registry/commit/42a43ca4296e912f456ed269999ab407b0e7b695

nmwsharp:
thanks for this! I went ahead and updated the included nanoflann at the same time, since it was an ancient version

PR #208, opened 2025-01-23
Update nanort version and include path.

jdumas:
This PR updates the include path of nanort in geometry central to facilitate using the upstream [nanort](https://github.com/lighttransport/nanort) repository. I.e. users should `#include <nanort.h>` and not `<nanort/nanort.h>`.

Additionally, since nanort in geometry central is 8 years old, I though we could update it as well, as geometry central code isn't compatible with the latest nanort API.

### DIFF
--- a/deps/CMakeLists.txt
+++ b/deps/CMakeLists.txt
@@ -79,7 +79,7 @@ else()
       FetchContent_GetProperties(eigen)
       if(NOT eigen_POPULATED)
           FetchContent_Populate(eigen) # NOTE: this form was deprecated in 3.30, we set CMP0169 to silence the warning
-          
+
           include(EigenChecker)
           eigen3checker(${eigen_SOURCE_DIR} 3.3)
       endif()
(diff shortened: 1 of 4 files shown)


nmwsharp:
Thank you for submitting this. I believe this is fully handled by the changes in https://github.com/nmwsharp/geometry-central/pull/242 incorporating these. If not please re-open!

There were several overlapping issues/PRs about the way nanoflann/nanort are handled.

I tried to unify & fix so now:

- Both nanort and nanoflann live on their standard paths
- Remove old backwards compatible code preserving old paths (this is a small breaking change, better than the confusion of multiple paths)
- Make nanort an private dependency similar to nanoflann make install/link in large codebases easier
- Create a small tracer class exposing nanort

jdumas:
Thanks. Out of curiosity, have you thought about using FetchContent/CPM to manage those dependencies? Then you could setup something like Dependabot/Renovate to keep them updated without breaking a sweat. It should be less work in the long run, and I'm happy to help if you want.

PR #215, opened 2025-05-26
Fix sign typo on "face Gaussian curvature" doc

DavidJourdan:
Hopefully the title is self-explanatory, there was a sign typo on the formula for computing the integrated Gaussian curvature on a face using the angle defect. The actual code doesn't have this typo, so this pull request only affects the docs.


nmwsharp:
good catch, thank you for pushing the fix

PR #216, opened 2025-05-27
Add build documentation for Bazel users

phaedon:


nmwsharp:
This is awesome! Thanks for doing it.

One question from my end -- is the bazel build registry something that now needs to be updated somehow with each new version of the library? Or will it pull new versions automatically after a tagged release?

PR #237, opened 2026-03-03
fix bug with flipEdgeNetwork created by piecewise Dijkstra

### LINKED ISSUE #209: Artifacts on the resulting intrinsic mesh
Really great repo. I've almost been able to implement a new modifier on blender thanks to it. 
Sadly tho I encountereed some major issues. 

Basically the algorithm goes like this. 
1. I take the mesh from blender 
2. I add some new vertices laying on the mesh through delaunay triangulation
3. I obtain the shortest path between these vertices
4. I convert it to gc mesh
5. I shorten the paths
6. I apply the delaunay triangulation from your lib on the resulting mesh
7. I mark the seams
8. I reconvert it to blender mesh and return the result to blender 

the initial vertices on the mesh are basically the path where to cut the mesh. 
The problem is that I have no way to get the edges of the strighten paths so i can mark them on blender

Now I'm doing it by getting the gc mesh, then using getPathPolyline3D and selecting the vertices on the mesh that are coincident by checking their distance, but I noticed that getPathPolyine3D even returns vertices that don't exist on the mesh

Also I notices some artifacts I couldn't be able to get rid. 
one is about some points on the mesh where the density becomes too high for no apparent reason
and the other issue is that getPathPolyline3D returns edges that make no sense.

Image
Image

Joel-Hecht:
Fixes #209

For some vertices A,B,C using shortestEdgePath() to connect vertex A to vertex B, and vertex B to vertex C would result in paths AB and BC that possibly contain one or more duplicate edges (i.e. one or more edges in AB has its twin in BC).  The flipout algorithm is not able to flip these 'back-and-forth' loops (or the edges immediately before/after them), leading to incorrect shortening of the input path.

PR adds an option (enabled as default) in constructFromPiecewiseDijkstraPath() to remove any subset of the resulting curve that starts and ends at the same vertex, which, for all halfedges in the subset, also contains their twins.

Only fixes the 'artifacts' problem in the linked issue.  Other 'problems' mentioned in the linked issue are fundamental misunderstandings of the algorithm.

Also adds a few contrived tests illustrating cases that would fail before, but now have the expected behavior.  These tests are not very interesting and can probably get deleted.

Possibly also fixes this issue noted in comment in wedgeIsClear() (?):
 ```TODO handle checks in case where the path lollipops out and back along a single edge```


nmwsharp:
Thank you for this! The problem with the "out-and-back" piecewise Dijkstra paths makes sense. I see how it could happen and agree this simple change helps a lot at least.

We can't apply it when marking vertices, but otherwise I think it should always be enabled. I pushed a commit to this PR that makes that tweak, as well as some other cleanup around the code style and test organization, but the algorithm should be unchanged.

I'm going to go ahead and merge, let me know if you spot any issues I introduced in translation.

PR #239, opened 2026-04-03
Fix: clear m_stop_vertices unconditionally in set_stop_conditions

selaytekgul:
Clear `m_stop_vertices` unconditionally in `set_stop_conditions` to prevent stale stop vertex accumulation when reusing `GeodesicAlgorithmExact` instances across multiple `propagate()` calls.


nmwsharp:
I haven't actually tested this, but it looks like a blatant bug to me so I'm going to go ahead and merge it.

Thank you for submitting this fix!

